Issue #7, opened 2022-07-14
Add __init__ example


chonkyh:
# Issue

Current README example only shows basic usage, running with __init__ for vyper contracts is currently unclear.
Propose adding an `examples` folder to store some examples that users can work with and also updating README with __init__ example.

PR #8, opened 2022-07-14
Add example for running with __init__

chonkyh:
- Add some examples for running with __init__
- created examples folder with ERC20.vy, simple.vy


chonkyh:
For issue #7

charles-cooper (commented):
looks good, two small issues --
- i think we can get rid of `examples/__init__.py`, unless there is some reason for it i am missing?
- there is already a `tests/ERC20.vy`. is there any way we can avoid copying the code?

chonkyh:
- removed `examples/__init__.py`
- propose moving `test/ERC20.vy` to `examples/ERC20.vy`

PR #19, opened 2022-08-30
Add forking example

imachonky:
Add forking mainnet example on README.md


imachonky:
@charles-cooper sounds good! wondering if having readthedocs would be good as there are more features getting added.

PR #48, opened 2022-12-16
Feat: profile calls

bout3fiddy:
Enables call profiling. User needs to set up their profiling tests similar to the following example:

```
@pytest.mark.parametrize(
    "a,b,c", [(42, 69, 1), (420, 690, 2), (42, 690, 3), (420, 69, 4)]
)
@pytest.mark.profile_calls("boa_contract.foo", "boa_contract.bar")
def test_populate_call_profile_property(boa_contract, a, b, c):

    boa_contract.foo(a, b)
    boa_contract.bar(a, b)

    assert boa_contract.call_profile
```

which generates:
```
tests/unitary/test_call_profiling/test_call_profile.py .....                                                                                     [100%] 
                                               
                  Call Profile                  
┏━━━━━━━━┳━━━━━━┳━━━━━━━━┳━━━━━━━┳━━━━━━┳━━━━━━┓
┃ Method ┃ Mean ┃ Median ┃ Stdev ┃ Min  ┃ Max  ┃
┡━━━━━━━━╇━━━━━━╇━━━━━━━━╇━━━━━━━╇━━━━━━╇━━━━━━┩
│    foo │ 188  │ 188.0  │ 0     │ 188  │ 188  │
│    bar │ 5275 │ 5275.0 │ 0     │ 5275 │ 5275 │
└────────┴──────┴────────┴───────┴──────┴──────┘
```

### DIFF
--- a/boa/test/plugin.py
+++ b/boa/test/plugin.py
@@ -32,21 +32,49 @@ def pytest_configure(config):
     pytest.call_profile = {}
 
 
+def _profile_calls(item):
+
+    profile_calls = item.get_closest_marker("profile_calls").args
+    if len(profile_calls) > 0:
+
+        combined_calls = {}
+        for call in profile_calls:
+
+            fixture_name = call.split(".")[0]
+            method_name = call.split(".")[1]
+
+            for name, fixture in item.funcargs.items():
+
+                if name == fixture_name:
+
+                    for d in (pytest.call_profile, fixture.call_profile):
+                        for key, value in d.items():
+                            if key != method_name:
+                                continue
+                            if key not in combined_calls.keys():
+                                combined_calls[key] = value
+                            else:
+                                combined_calls[key].extend(value)
+
+        pytest.call_profile = combined_calls
+
+
 @pytest.hookimpl(hookwrapper=True)
 def pytest_runtest_call(item: pytest.Item) -> Generator:
 
-    ignore_isolation = item.get_closest_marker("ignore_isolation")
-    profile_calls = item.get_closest_marker("profile_calls")
-    isolate = not (ignore_isolation and profile_calls)
+    ignore_isolation = item.get_closest_marker("ignore_isolation") is not None
 
-    if isolate:
-
+    if not ignore_isolation:
         with boa.env.anchor():
             yield
-
     else:
-
         yield
+
+    profile_test = item.get_closest_marker("profile_calls") is not None
+    if not profile_test:
+        return
+
+    _profile_calls(item)
 
 
 def pytest_sessionfinish(session, exitstatus):
--- a/boa/test/plugin.py
+++ b/boa/test/plugin.py
@@ -27,13 +27,74 @@ hypothesis.core.HypothesisHandle.__init__ = _HypothesisHandle__init__  # type: i
 
 def pytest_configure(config):
     config.addinivalue_line("markers", "ignore_isolation")
+    config.addinivalue_line("markers", "profile_calls")
+
+    pytest.call_profile = {}
 
 
 @pytest.hookimpl(hookwrapper=True)
 def pytest_runtest_call(item: pytest.Item) -> Generator:
 
-    if not item.get_closest_marker("ignore_isolation"):
+    ignore_isolation = item.get_closest_marker("ignore_isolation")
+    profile_calls = item.get_closest_marker("profile_calls")
+    isolate = not (ignore_isolation and profile_calls)
+
+    if isolate:
+
         with boa.env.anchor():
             yield
+
     else:
+
         yield
+
+
+def pytest_sessionfinish(session, exitstatus):
+
+    if not pytest.call_profile:
+        pass
+
+    import statistics
+    import sys
+
+    from rich.console import Console
+    from rich.table import Table
+
+    # generate means, stds for each
+    call_profiles = {}
+    for key, gas_used in pytest.call_profile.items():
+        call_profile = {}
+
+        call_profile["mean"] = statistics.mean(gas_used)
+        call_profile["median"] = statistics.median(gas_used)
+        call_profile["min"] = min(gas_used)
+        call_profile["max"] = max(gas_used)
+        if len(gas_used) == 1:
+            call_profile["stdev"] = 0
+        else:
+            call_profile["stdev"] = int(round(statistics.stdev(gas_used), 2))
+
+        call_profiles[key] = call_profile
+
+    # print rich table
+    table = Table(title="\nCall Profile")
+    table.add_column("Method", justify="right", style="cyan", no_wrap=True)
+    table.add_column("Mean", style="magenta")
+    table.add_column("Median", style="magenta")
+    table.add_column("Stdev", style="magenta")
+    table.add_column("Min", style="magenta")
+    table.add_column("Max", style="magenta")
+
+    for key in call_profiles.keys():
+        profile = call_profiles[key]
+        table.add_row(
+            key,
+            str(profile["mean"]),
+            str(profile["median"]),
+            str(profile["stdev"]),
+            str(profile["min"]),
+            str(profile["max"]),
+        )
+
+    console = Console(file=sys.stdout)
+    console.print(table)
--- a/boa/test/plugin.py
+++ b/boa/test/plugin.py
@@ -32,6 +32,23 @@ def pytest_configure(config):
     pytest.call_profile = {}
 
 
+def _enable_call_profiling(item):
+    profile_test = item.get_closest_marker("profile_calls") is not None
+    if not profile_test:
+        return
+
+    profile_calls = item.get_closest_marker("profile_calls").args
+    if len(profile_calls) > 0:
+
+        for call in profile_calls:
+            fixture_name = call.split(".")[0]
+            for name, fixture in item.funcargs.items():
+
+                if name == fixture_name:
+
+                    fixture.profile_calls = True
+
+
 def _profile_calls(item):
 
     profile_calls = item.get_closest_marker("profile_calls").args
@@ -63,6 +80,7 @@ def _profile_calls(item):
 def pytest_runtest_call(item: pytest.Item) -> Generator:
 
     ignore_isolation = item.get_closest_marker("ignore_isolation") is not None
+    _enable_call_profiling(item)
 
     if not ignore_isolation:
         with boa.env.anchor():
--- a/boa/vyper/contract.py
+++ b/boa/vyper/contract.py
@@ -420,6 +420,10 @@ class VyperContract(_BaseContract):
         if not skip_init:
             self._run_init(*args)
 
+        # call profiling
+        self.profile_calls = False
+        self.call_profile = {}
+
     def _run_init(self, *args):
         encoded_args = b""
 
@@ -465,15 +469,18 @@ class VyperContract(_BaseContract):
     def ast_map(self):
         return ast_map_of(self.compiler_data.vyper_module)
 
-    def debug_frame(self, computation=None):
-        if computation is None:
-            computation = self._computation
-
+    def _get_fn_from_computation(self, computation):
         node = self.find_source_of(computation.code)
         if node is None:
             return None
 
-        fn = node.get_ancestor(vy_ast.FunctionDef)
+        return node.get_ancestor(vy_ast.FunctionDef)
+
+    def debug_frame(self, computation=None):
+        if computation is None:
+            computation = self._computation
+
+        fn = self._get_fn_from_computation(computation)
 
         frame_info = self.compiler_data.function_signatures[fn.name].frame_info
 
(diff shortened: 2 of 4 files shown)


boa/test/plugin.py
```diff
@@ -27,13 +27,129 @@ def f(*args, **kwargs):
...
+    profile_calls = item.get_closest_marker("profile_calls").args
+    if len(profile_calls) > 0:
+
+        for call in profile_calls:
+            fixture_name = call.split(".")[0]
+            for name, fixture in item.funcargs.items():
+
+                if name == fixture_name:
+
+                    fixture.profile_calls = True
```
charles-cooper, line 57: this has a couple problems. first of all, `profile_calls` is not unset at the end of the item. second of all, the approach is brittle. it relies on parsing the pytest arguments at runtime. what if there is a collision between fixture names and called function names?

it's going to be much cleaner if `enable_call_profiling` sets a flag globally (or on `boa.env`) which can be checked by all contracts at execution, and it's easier to reason about the state of the flag (set/unset) and has less dependency to set up and teardown, since the flag is either set or unset.

charles-cooper (commented):
i see this is a draft, but i left some feedback on the approach. i think it will be much cleaner to reason about if the `profiling` flag is set in a more global place, rather than on a per-contract basis. the current approach does not clear the flag, so it has a memory leak because all tests will end up profiling (and not clearing the profiling data).

i think it might also be cleaner to set it up so that profiling happens "lazily", at the end of the entire test suite, rather than right after each test. this is going to be more performant, since the profiler path will only be taken once as a batch action at the end of the test suite, and also feel more performant (to the user), since profiling won't happen during the execution of the tests themselves, and we may even be able to disable profile report generation in a lot of cases (ex. if the test suite fails). lastly, it will be more general, since if we want to add other instrumentation later, we will not need to do it at the time of the test, we can just calculate it "after the fact". so the approach i suggest is, to just append computations an Environment-scoped variable, keeping track of all computation objects that have happened during profiling tests. then at the end of the test suite, we can iterate over them and build a profile object. there may be some undesirable edge cases which have less good reasonability properties than the "eager" way, but besides contract code changing mid-test suite, i'm having trouble thinking of good ones.

bout3fiddy:
Following comments from @charles-cooper, I've introduced changes that allow lazily profiling calls. Usage:

```
@pytest.mark.profile_calls
def test_something(contract):
    contract.foo()
```

bout3fiddy:
not sure how to make this table look pretty in reStructuredText:

image

boa/test/plugin.py
```diff
@@ -15,8 +16,15 @@ def _HypothesisHandle__init__(self, *args, **kwargs):
...
+    has_profiling_flag = False
+    if "pytestmark" in t.__dict__:
+        for mark in t.pytestmark:
+            if mark.name == "profile_calls":
+                has_profiling_flag = True
+                break
+
     def f(*args, **kwargs):
-        with boa.env.anchor():
+        with boa.env.anchor(), boa.env.store_call_profile(has_profiling_flag):
```
charles-cooper, line 27: i don't think this is required to do on every test run, since the pytest machinery should already take care of it.

boa/environment.py
```diff
@@ -266,6 +266,9 @@ def __init__(self):
 
         self._contracts = {}
 
+        self.profile_calls = False
+        self.profiled_calls = {}
```
charles-cooper, line 270: both of these should be private

boa/environment.py
```diff
@@ -341,6 +344,14 @@ def gas_meter_class(self, cls):
...
         self.set_gas_meter_class(ProfilingGasMeter)
 
+    @contextlib.contextmanager
+    def store_call_profile(self, enable: bool = True) -> None:
+        try:
+            self.profile_calls = enable
+            yield
+        finally:
+            self.profile_calls = False
+
```
charles-cooper, line 354: I'm not convinced that we need to expose this on the environment object. I can't think of why a user would want to use it -- it's pretty much only there for testing purposes at this time. instead of having this method, we should just inline this code in `pytest_runtest_call`

boa/vyper/contract.py
```diff
@@ -585,6 +588,20 @@ def marshal_to_python(self, computation, vyper_typ):
...
+        if self.env.profile_calls:
+            try:
+                fn_name = self._get_fn_from_computation(computation).name
+            except AttributeError:
+                # TODO: https://github.com/vyperlang/vyper/issues/3200  # noqa: E501
+                fn_name = "unnamed"
+            fn_name = self.compiler_data.contract_name + "." + fn_name
+            gas_used = computation.get_gas_used()
+            if fn_name not in self.env.profiled_calls.keys():
+                self.env.profiled_calls[fn_name] = [gas_used]
```
charles-cooper, line 601: there is a problem here if two contracts have the same name and have a function with the same name.

pyproject.toml
```diff
@@ -21,17 +21,15 @@ dependencies = [
     "eth-typing",
     "hypothesis",
     "pytest",
-#'eth-stdlib; python_version >= "3.10"',
+    "rich",
```
charles-cooper, line 24: should this be an extra?

boa/environment.py
bout3fiddy: Well they could use this for their own statistics. This just offers a way to capture gas used. in boa.env.profiled_calls dict. Nothing more really.

boa/test/plugin.py
bout3fiddy: it does not do so for hypothesis (which is why it is here).

pyproject.toml
bout3fiddy: I think it's fine. In fact, we should consider using rich in more places (because it is such an awesome module for prettifying console outputs).

It doesn't really bloat up boa, and is only imported at the end of the pytest session.

boa/test/plugin.py
```diff
@@ -1,22 +1,40 @@
...
 _old_init = hypothesis.core.HypothesisHandle.__init__
 
 
+@contextlib.contextmanager
+def _enable_profiling(flag: bool):
+    try:
+        boa.env._profile_calls = flag
+        yield
+    finally:
+        boa.env._profile_calls = False
```
charles-cooper, line 21: this has an interesting effect of disabling `_profile_calls` even if it was set to `True` before `_enable_profiling()` was called.

boa/profiling.py
```diff
@@ -133,3 +137,125 @@ def summary(
...
+
+
+# simple dataclass for fn
+SelectorInfo = namedtuple("SelectorInfo", ["fn_name", "contract_name", "address"])
+
+
+# cache gas_used for all computation (including children)
+def cache_gas_used_for_computation(computation, env):
+
+    contract = env.lookup_contract(computation.msg.storage_address)
```
charles-cooper, line 149: this could fail for a black box contract

charles-cooper: it might be useful to use the `Datum` abstraction earlier in this file

boa/profiling.py
```diff
@@ -133,3 +137,125 @@ def summary(
...
+        # https://github.com/vyperlang/vyper/pull/3202
+        fn_name = "unnamed"
+
+    fn = SelectorInfo(
+        fn_name=fn_name,
+        contract_name=contract.compiler_data.contract_name,
+        address=to_checksum_address(contract.address),
+    )
+
+    gas_used = computation.get_gas_used()
```
charles-cooper, line 164: doesn't handle gas refund -- check `Datum` abstraction as mentioned here https://github.com/vyperlang/titanoboa/pull/48/files/4349f08f1c5964592ee1800fd3a37e5d04a597b5#r1051643690

boa/test/plugin.py
bout3fiddy: renamed it to `_toggle_profiling`

boa/vyper/contract.py
```diff
@@ -465,15 +466,17 @@ def at(self, *args, **kwargs):
     def ast_map(self):
         return ast_map_of(self.compiler_data.vyper_module)
 
+    def _get_fn_from_computation(self, computation):
+        node = self.find_source_of(computation.code)
+        if isinstance(node, vy_ast.FunctionDef):
```
charles-cooper, line 471: i think this breaks the previous None check?

bout3fiddy: It will not, since that is in the profiling section https://github.com/vyperlang/titanoboa/blob/dae534980793c36704f97f53bdecc28b8bb591ee/boa/profiling.py#L181-L188

boa/profiling.py
```diff
@@ -1,8 +1,20 @@
...
+
+from boa.environment import Env
 from boa.vyper.ast_utils import get_line
 
+# simple dataclass for fn
+CallInfo = namedtuple("SelectorInfo", ["fn_name", "contract_name", "address"])
+LineInfo = namedtuple(
+    "LineGasUsedInfo", ["address", "contract_name", "lineno", "line_src"]
+)
+
```
charles-cooper, line 17: why not just use actual dataclasses here?

boa/test/plugin.py
```diff
@@ -1,15 +1,27 @@
...
+from boa.profiling import get_call_profile_table, get_line_profile_table
+from boa.vm.gas_meters import ProfilingGasMeter
 
 # monkey patch HypothesisHandle. this fixes underlying isolation for
 # hypothesis.given() and also hypothesis stateful functionality.
 _old_init = hypothesis.core.HypothesisHandle.__init__
 
 
+@contextlib.contextmanager
+def _toggle_profiling(yes: bool = False):
```
charles-cooper, line 17: this function is confusing. if the flag is False, it doesn't disable profiling. also, it's only used in one place - maybe just inline it there?

boa/profiling.py
```diff
@@ -102,34 +117,267 @@ def merge(self, other: "LineProfile") -> None:
...
+        else:
+            env._cached_line_profiles[line].append(gas_used)
+
+    # ------------------------------------------------------------
+
+    # recursion for child computations
+    for _computation in computation.children:
+        child_contract = env.lookup_contract(_computation.msg.code_address)
+        # ignore black box contracts
+        if child_contract:
```
charles-cooper, line 216:
```suggestion
        if child_contract is not None:
```

boa/profiling.py
```diff
@@ -102,34 +117,267 @@ def merge(self, other: "LineProfile") -> None:
...
+        contract_uid = (lp.contract_name, lp.address)
+
+        # add spaces so numbers take up equal space
+        lineno = str(lp.lineno)
+        if lp.lineno < 10:
+            lineno = "   " + lineno
+        if 10 <= lp.lineno < 100:
+            lineno = "  " + lineno
+        if 100 <= lp.lineno < 1000:
+            lineno = " " + lineno
```
charles-cooper, line 324: this whole block can just be a call to `rjust`

boa/profiling.py
```diff
@@ -102,34 +117,267 @@ def merge(self, other: "LineProfile") -> None:
...
+
+        gas_data = (lineno + ": " + lp.line_src, gas_data)
+        if contract_uid not in contracts:
+            contracts[contract_uid] = [gas_data]
+        else:
+            contracts[contract_uid].append(gas_data)
```
charles-cooper, line 330: small thing but this will be more concise with a call to `setdefault()`:
```suggestion
        s = contracts.setdefault(contract_uid, [])
        s.append(gas_data)
```

boa/profiling.py
```diff
@@ -102,34 +117,267 @@ def merge(self, other: "LineProfile") -> None:
...
+                    len(gas_used),
+                    int(statistics.mean(gas_used)),
+                    int(statistics.median(gas_used)),
+                    stdev,
+                    min(gas_used),
+                    max(gas_used),
+                ]
+            )
+
+        l_profile = sorted(l_profile, key=lambda x: int(x[3]), reverse=True)
```
charles-cooper, line 359: comment what `x[3]` is please

boa/profiling.py
```diff
@@ -102,34 +117,267 @@ def merge(self, other: "LineProfile") -> None:
...
+                [
+                    contract_name,
+                    code,
+                    len(gas_used),
+                    int(statistics.mean(gas_used)),
+                    int(statistics.median(gas_used)),
+                    stdev,
+                    min(gas_used),
+                    max(gas_used),
+                ]
```
charles-cooper, line 356: style: unless the item should be mutable, make it a tuple

charles-cooper: since this code is similar to the long block in `get_call_profile_table()`, can we factor it into a Stats object?

charles-cooper (commented):
looking good, some style changes requested. also there are long blocks which are just moving around clumps of stats results, can you please refactor those into a class and maybe some convenience methods for working with it? you can get `list()` behavior by implementing iterator or bound indexing on the object: https://stackoverflow.com/a/30554680

PR #51, opened 2022-12-28
Handle empty hex strings ('0x') returned by latest ganache versions

benber86:
When using boa with a ganache (v7.7.2) local fork, I ran into the following error when trying to retrieve a value:

image

This happens because recent version of ganache return "0x" for empty values instead of "0x0" or "0x0000000000000000000000000000000000000000000000000000000000000000". 

This fix handles this edge case


bout3fiddy:
LGTM

charles-cooper (approved):

PR #55, opened 2022-12-28
Extract reason comment from multiline statement

### LINKED ISSUE #52: bug: error parsing reason comment on multi-line statements

benber86:
Fixes #52 
I've modified `DevReason` to remove `line_no` in the `at` method and simply pass the multiline source code to extract comments from. If better to keep ability to localize error with line numbers, could add an `end_line` instead and extract the relevant lines accordingly.


boa/vyper/ast_utils.py
```diff
@@ -6,9 +6,13 @@
 from vyper.codegen.core import getpos
 
 
-def get_line(source_code: str, lineno: int) -> str:
+def get_block(source_code: str, lineno: int, end_lineno: int) -> str:
```
bout3fiddy, line 9: This is introducing a breaking change for the line profiler: https://github.com/benber86/titanoboa/blob/c537ea1d9f1d61bf56288ab24ae1d99c77a3c3a8/boa/profiling.py#L121

So you'll need to address this. I think the block idea is good because line profiler can look quite ugly if the lines are multi-line.

benber86: I didn't remove `get_line`, just moved it two lines down and refactored to use `get_block`:

```
def get_line(source_code: str, lineno: int) -> str:
    return get_block(source_code, lineno, lineno)
```

PR #60, opened 2023-01-23
Adds a dictionary of event args/data to Event type for easier access to variables

kibagateaux:
## Description
- Updates `vyper/contracts.py` and `vyper/events.py` to construct a dict from event data when decoding and return with other event data
- Allows access to event data with event variable names e.g. `contract.get_logs()[0].args_map['my_event_variable']` 

# Testing
 Currently no testing for events. Would appreciate advice on how you guys want that setup.


bout3fiddy:
It’s missing a pre-commit run. Also some simple tests using a mock contract will help quite a bit! Something like what you have in your repo, but minimal.

PR #178, opened 2024-03-01
fix: unit and integration tests

DanielSchiavini:
image

### DIFF
--- a/boa/interpret.py
+++ b/boa/interpret.py
@@ -1,7 +1,7 @@
-import importlib
 import json
 import sys
 import textwrap
+from importlib.abc import MetaPathFinder
 from importlib.machinery import SourceFileLoader
 from importlib.util import spec_from_loader
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


charles-cooper (approved):
looks good. can we swap the order of BoaLoader and BoaImporter so it follows the logic / execution flow?

boa/interpret.py
```diff
@@ -26,39 +28,30 @@
...
-    def load_module(self, fullname):
-        if fullname in sys.modules:
-            return sys.modules[fullname]
 
-        # import system should guarantee this, but be paranoid
-        if fullname not in self._path_lookup:
-            raise ImportError(f"invariant violated: no lookup for {fullname}")
+class BoaLoader(SourceFileLoader):
+    def get_code(self, fullname):
+        return ""
```
charles-cooper, line 45: per https://docs.python.org/3/library/importlib.html#importlib.abc.InspectLoader.get_code, this should return None

DanielSchiavini: image

charles-cooper: nvm, discussed offline -- apparently returning `None` triggers an `ImportError`.

charles-cooper: let's add a comment though
```suggestion
        # importlib docs say to return None, but that triggers an `ImportError`
        return ""
```

PR #182, opened 2024-03-04
Disable forking network env

DanielSchiavini:
### What I did
- Changed interface from `boa.env.fork` to `boa.fork`

### How I did it
- Raise error if `fork` is called on a `NetworkEnv`

### How to verify it
- boa.set_network_env(url)
- boa.env.fork should now raise an error

- Changed interface from `boa.env.fork` to `boa.fork`

### Cute Animal Picture
image


boa/__init__.py
```diff
@@ -58,6 +58,16 @@ def set_network_env(url):
     set_env(NetworkEnv.from_url(url))
 
 
+def fork(*args, **kwargs):
+    """Fork the current environment to use a custom network URL"""
+    e = Env()
+    if env.__name__ == "BrowserEnv":
```
DanielSchiavini, line 64: this does not work, we need to find a better solution

PR #194, opened 2024-04-16
docs: access private/external members

AlbertoCentonze:
### What I did
Added documentation on how to access variables/functions that are not exposed by default.

### How to verify it
Read The Fine Manual ;)

Same as title

### Cute Animal Picture
image


charles-cooper (commented):
this is helpful! but i think it might belong in a section besides "testing"

charles-cooper (approved):
merging, but we should reorganize later

PR #203, opened 2024-05-06
fix: local variable 'snapshot_id' referenced before assignment

DanielSchiavini:
### What I did
Fix UnboundLocalError: local variable 'snapshot_id' referenced before assignment

This was caused by a typo, the property `block_id` was removed in #134. Another typo broke the time travel tests.

### How I did it
Used the correct `block_number` property

### How to verify it
Unit tests should pass

Fix UnboundLocalError: local variable 'snapshot_id' referenced before assignment

### Cute Animal Picture

image


charles-cooper (approved):
thank you, lgtm

PR #237, opened 2024-05-27
improve profiling output

bout3fiddy:
update view; arrange profiling table on median and not mean (so users can ignore storage slot writes and look at computations instead)

### What I did

call profile column looks like the following:

image

1. one less column (no address column. it's all in column 1)
2. paths are shorter
3. values are arranged based on median gas costs and not average gas costs

profiling table looks like:

image

1. paths are shorter
2. values are arranged based on median gas costs

arranging by median means storage reads and writes do not populate the top of the table. this is much better when it comes to looking for computations to optimise.

### How I did it

very simple code changes

### How to verify it

run a profiling test

improve profiling table view and arrange data by median gas costs

### Cute Animal Picture

image

### DIFF
--- a/boa/profiling.py
+++ b/boa/profiling.py
@@ -11,12 +11,13 @@ from boa.contracts.vyper.ast_utils import get_fn_name_from_lineno, get_line
 from boa.environment import Env
 
 
-def safe_relpath(path):
+def _safe_relpath(path):
     try:
-        relpath = os.path.relpath(path)
-    except ValueError:  # for Windows, if path is not on os.curdir
-        relpath = ""
-    return relpath
+        return os.path.relpath(path)
+    except ValueError:
+        # on Windows, if path and curdir are on different drives, an exception
+        # can be thrown
+        return path
 
 
 @dataclass(unsafe_hash=True)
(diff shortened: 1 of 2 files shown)


DanielSchiavini (approved):

boa/profiling.py
```diff
@@ -306,38 +302,45 @@ def get_call_profile_table(env: Env) -> Table:
...
             cname = ""
-            caddr = ""
-            if c == 0:
-                cname = profile.contract_name
-                caddr = address
             fn_name = profile.fn_name
-            table.add_row(cname, caddr, fn_name, *stats.net_gas_stats.get_str_repr())
+            stats = list(stats.net_gas_stats.get_str_repr())
+            if c == 0:
+                relpath = os.path.relpath(profile.contract_name)
```
charles-cooper, line 334: what if call to relpath fails?

boa/profiling.py
```diff
@@ -359,10 +362,10 @@ def get_line_profile_table(env: Env) -> Table:
 
     table = _create_table(for_line_profile=True)
     for (contract_name, contract_address), fn_data in contracts.items():
-        contract_file_path = os.path.split(contract_name)
+        relpath = os.path.relpath(contract_name)
```
charles-cooper, line 365: ditto

boa/profiling.py
bout3fiddy: it cannot fail as long as contract_name isnt `""`

charles-cooper: On windows it can fail if the two paths are on different drives
Screenshot_20240717_093152_Brave

PR #259, opened 2024-07-19
Fix KeyError when error happens in library

### LINKED ISSUE #252: `boa.reverts()` fails on assertion failure in a nested module
Can reproduce from here:

https://github.com/curvefi/autobribe/blob/4909778e459fa5a9646bf2f7346e975280a0ae29/test/unitary/test_incentives_manager.py#L69-L70

DanielSchiavini:
Fixes #252

### What I did
- Fixed KeyError

### How I did it
- By ignoring the debug frames of library errors

### How to verify it
- Test is included

N/A

### Cute Animal Picture
image

### DIFF
--- a/boa/contracts/vyper/vyper_contract.py
+++ b/boa/contracts/vyper/vyper_contract.py
@@ -651,11 +651,9 @@ class VyperContract(_BaseVyperContract):
             # TODO: figure out why fn is None.
             return None
 
-        signatures = self.compiler_data.function_signatures
-        if fn.name not in signatures:
-            return None  # todo: recurse into imported libraries
+        fn_t = fn._metadata["func_type"]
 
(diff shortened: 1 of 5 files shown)


boa/contracts/vyper/vyper_contract.py
```diff
@@ -651,7 +651,11 @@ def debug_frame(self, computation=None):
             # TODO: figure out why fn is None.
             return None
 
-        frame_info = self.compiler_data.function_signatures[fn.name]._ir_info.frame_info
+        signatures = self.compiler_data.function_signatures
+        if fn.name not in signatures:
```
DanielSchiavini, line 655: @charles-cooper is there an easy way to get the signatures of imported modules or to get the errored module from the ast node?

PR #270, opened 2024-08-02
docs: correct network mode usage

ThatOneCalculator:
### What I did

Adjusted the documentation to reflect the proper way to use network mode

### How I did it

Tried what was in the readme, failed, looked through the titanoboa codebase, looked at the eth_account documentation, and then tested this

### How to verify it

Run that very example

docs: correct network mode usage

### Cute Animal Picture

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,10 +199,10 @@ Cast current deployed addresses to vyper contract
 
 ```python
 >>> import boa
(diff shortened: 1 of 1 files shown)


charles-cooper (approved):
Thanks!

PR #302, opened 2024-09-06
fix: Etherscan retry for boa 0.1

DanielSchiavini:
### What I did
- Applied the fix of https://github.com/vyperlang/titanoboa/pull/287 to boa 0.1 so that users can use boa with vyper 0.3
- Fixed linting issues

### Cute Animal Picture
image


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "titanoboa"
-version = "0.1.10b1"
```
DanielSchiavini, line 3: looks like the v0.1.10 tag had the wrong version: https://github.com/vyperlang/titanoboa/blob/v0.1.10/pyproject.toml#L3

PR #320, opened 2024-09-30
fix[verify]: check whether the contract is really verified

DanielSchiavini:
### What I did
@AlbertoCentonze tried to verfiy a contract with boa. The contract could be found in `/smart-contracts` API call, but misses the `is_verified` field. This PR should add a check to make sure the contract is actually verified.
I'm trying to contact Blockscout devs to check what's going on in this case, given the verification via UI worked fine.

### Cute Animal Picture
image

### DIFF
--- a/boa/verifiers.py
+++ b/boa/verifiers.py
@@ -104,7 +104,7 @@ class Blockscout:
         if response.status_code in self.retry_http_codes:
             return False
         response.raise_for_status()
-        return True
+        return response.json().get("is_verified", False)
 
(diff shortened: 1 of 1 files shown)


rimrakhimov:
The thing is that `/api/v2/smart-contracts/{address}` returns errors only in case if the address is not a contract at all. Otherwise, when the contract is not verified the endpoint will just return non-verification specific data. So, the approach suggested here seems quite reasonable

PR #391, opened 2025-03-27
feat: improve load_partial with base_path argument

s3bc40:
### Cute Animal Picture


s3bc40:
Draft the PR since the issue might be more complex to make VVM vyper to take into account the mox `lib` path system.
```py
from snekmate.auth import ownable as ow # does not work in VVM
from lib.pypi.snekmate.auth import ownable as ow # it works
```
Maybe something to do in the VVM project or Mox side...

PR #421, opened 2025-08-17
fix: simplify ABI and VVM error output

AlbertoCentonze:
### What I did

Simplify ABI and VVM error output to show the call tree once, while retaining native Vyper source-level diagnostics.

ABI/VVM reverts previously repeated the same call information and included raw VM arguments after the call tree. They now display the concise tree. Native Vyper failures raise `VyperError`, a subclass of `BoaError`, and continue to include source, locals, storage, and compiler diagnostics. Existing `except boa.BoaError` handlers and programmatic `call_trace`, `stack_trace`, and exception arguments remain supported.

### How I did it

- Share the call-tree formatting between `BoaError` and `VyperError`.
- Preserve the missing-bytecode warning in ABI call traces, addressing the regression noted in the earlier discussion.
- Keep revert reasons visible for unknown sources, including failed VVM constructors, and retain non-revert VM failures such as out-of-gas. Undecodable revert data remains visible as hex.
- Resolve conflicts against current `master`, remove the unrelated ignore-file change, update the error documentation, and add regression tests.

Stack traces are still constructed eagerly; this changes their display, not their availability.

### How to verify it

```sh
pytest -q tests/unitary/test_error_formatting.py tests/unitary/test_reverts.py tests/unitary/contracts/abi/test_abi.py tests/unitary/contracts/vvm/test_vvm.py
pytest -q tests/integration/network/anvil
```

Local validation on Python 3.12 / Vyper 0.4.3:

- Focused error, revert, ABI, and VVM suite: 65 passed, including 10 new regression cases.
- Anvil integration: 11 passed.
- Pre-commit checks for all changed files and package-wide mypy: passed.
- The full unit run had 458 passed, 4 failed, 1 skipped, and repeated collection errors from `test_isolation.py`; coverage was 49.86% against the unchanged 78% requirement. Failures concern pytest fixture marks, Jupyter compatibility, macOS coverage path normalization, and a timing-sensitive environment timestamp test. The fixture-mark and SignedMessage failures were also reproduced on unchanged `master`.

The base branch already has [failing unit CI](https://github.com/vyperlang/titanoboa/actions/runs/27811948658/job/82304111394) and unrelated full-tree lint failures. Those are being addressed separately; this PR does not skip tests, lower coverage requirements, or change CI access controls.

ABI and VVM errors are easier to read, with less redundant output and preserved diagnostic warnings and failure reasons. Native Vyper errors retain detailed source information.


AlbertoCentonze:
A new test is failing because a warning disappeared, which was to be expected given my changes. However that warning has been a lifesaver multiple times for me, so I should probably find a way to add it into call trees. Possibly in a separate PR to keep this one clean.

PR #454, opened 2026-06-16
Add eval and injection for VVM contracts

AlbertoCentonze:
## Summary

- keep VVM compiler/source context so deployed VVM contracts can expose debug helpers
- return a VVMContract from VVM deployers and add inject_function / eval support
- allow ABI functions to execute with temporary override bytecode


charles-cooper:
not sure why these features weren't working before but i'm assuming you have been testing the patch on your code!

PR #457, opened 2026-09-05
Fix CI coverage collection and baseline test failures

AlbertoCentonze:
Collect boa's pytest-plugin imports before pytest starts, then combine those measurements with pytest-cov's xdist worker data before applying the existing 78% coverage gate. Replace the obsolete coverage environment variables with an explicit configuration.

Update the isolation and signing tests for current pytest and eth-account, and make shared-memory, coverage-path and timestamp assertions work across platforms and longer test runs. Apply the existing Black/isort rules and remove unused imports.

Retry transient SQLite journal-mode initialization locks with a bounded deadline. Add regressions for lock recovery, timeout reporting and immediate propagation of unrelated SQL errors.

### DIFF
--- a/boa/util/sqlitedb.py
+++ b/boa/util/sqlitedb.py
@@ -99,7 +99,18 @@ class SqliteCache(BaseDB):
 
         # executescript does not work in multiprocess environment
         for cmd in self._PRAGMA_CMDS:
-            self._cursor.execute(cmd)
+            # Switching journal mode can contend with another worker before
+            # we can acquire a transaction lock. Keep retries bounded and
+            # propagate errors unrelated to contention.
+            deadline = time.monotonic() + 10
+            while True:
+                try:
+                    self._cursor.execute(cmd)
+                    break
+                except sqlite3.OperationalError as e:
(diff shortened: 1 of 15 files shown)


boa/util/sqlitedb.py
```diff
@@ -99,7 +99,18 @@ def __init__(self, db_path: Path | str, ttl: float = _ONE_MONTH) -> None:
 
         # executescript does not work in multiprocess environment
         for cmd in self._PRAGMA_CMDS:
-            self._cursor.execute(cmd)
+            # Switching journal mode can contend with another worker before
+            # we can acquire a transaction lock. Keep retries bounded and
+            # propagate errors unrelated to contention.
+            deadline = time.monotonic() + 10
+            while True:
```
charles-cooper, line 106: is there a more graceful way to acquire the lock? or what do other parts of the codebase do?

charles-cooper (approved):
approved in principle, there are a bunch of lint changes i guess which i'm not sure they sneaked in or are getting it compliant with latest linter configured in the CI

